PR #79, opened 2026-09-15
The baseline contract asserts a checkout exists, not which release it is (#78)

### LINKED ISSUE #78: The baseline contract asserts an action version where it means a structure
## Context

Found by the dependabot config from #74 on its **first** automated PR. `#77` bumped the actions group and `scripts/test-reusable-pr-baseline.mjs` went red. Diagnosed by @neo-opus-ada at @tobiu's request: the bump is correct and the test is wrong.

```
AssertionError: canonical reusable workflow violates its own contract
  actual:   [ 'missing caller checkout' ]
  expected: []
```

The contract has been version-coupled since it was written, and nothing surfaced it because nothing had bumped those actions. The first automated bump found it in seconds.

## The Problem

`scripts/test-reusable-pr-baseline.mjs:64`:

```js
['caller checkout', /uses: actions\/checkout@v4/, skillsJob],
```

The contract means *"the `skills-materialized` job checks out the caller's repository."* It expresses that as *"matches `checkout@**v4**`"*. Dependabot moved `v4 → v7`, the regex misses, and **a job that still has its checkout is reported as having lost it.**

A version pin standing in for a structural assertion, so every legitimate upgrade reads as the structure disappearing. Same defect class the org has hit repeatedly this week: a guard asking a **narrower** question than the condition it means to test, failing in the direction that looks like a real finding.

**Five further spots couple to a version** — one more than the initial report, which missed `:391`:

```
:265  mutation replaces the literal '- uses: actions/checkout@v4'
:316  fixture string embeds 'actions/setup-node@v4'
:317  fixture string embeds 'actions/setup-node@v4'
:391  mutation replaces a block containing 'actions/github-script@v7'
:407  mutation inserts 'actions/checkout@v4'
```

## The Architectural Reality

**These five are noisy, not dangerous, and the reason matters** — the initial report had it the other way round.

`expectMutationFailure` already carries the guard that makes a dead mutation impossible to miss:

```js
// scripts/test-reusable-pr-baseline.mjs:237
assert.notEqual(mutated, source, `${label}: fixture mutation changed nothing`);
```

So a `.replace()` whose target no longer exists does **not** pass for the wrong reason — line 237 fails loudly, naming the arm and saying the fixture mutation changed nothing. The recommendation to *"assert the replacement actually changed the source"* is already implemented. Whoever wrote that assertion anticipated exactly this, and it is why the next bump produces a clear failure instead of a silent hole.

That splits the work honestly: **`:64` is a correctness defect** (a false finding on a correct workflow), and the five `.replace()` sites are **upgrade friction** (every actions bump reds the suite until someone retypes a literal). Both worth fixing; only one is a false report.

## The Fix

**`:64` — assert the structure, not the release:**

```js
['caller checkout', /uses: actions\/checkout@v\d+/, skillsJob],
```

**The five `.replace()` sites — anchor on something version-free.** Prefer a target that carries the meaning (`- uses: actions/checkout@`, the `with:` block, the step `name:`) so a bump cannot invalidate it. Line 237 stays exactly as it is: it is the reason a mistake here is loud, and it should not be weakened into a warning.

## Decision Record impact

`none`. A test asserts what it already meant to assert.

## Acceptance Criteria

- [ ] `:64` matches any major — `checkout@v\d+` or equivalent — and the canonical workflow passes its own contract at the current action versions.
- [ ] `test-reusable-pr-baseline.mjs` exits 0 against `#77`'s bumped workflow (`checkout@v7`), which is the live reproducer.
- [ ] The five `.replace()` targets no longer embed a major version; each anchors on a version-free substring.
- [ ] `assert.notEqual(mutated, source, …)` at `:237` is **unchanged** — an AC rather than a note, because removing it while making the targets "safer" would trade a loud failure for a silent one.
- [ ] A regression arm proves `:64` accepts a bumped major **and still rejects a missing checkout** — the second half is what stops the fix from becoming a regex that matches anything.
- [ ] `node scripts/test-reusable-pr-baseline.mjs` and `node scripts/lint-skill-corpus.mjs --base origin/dev` both green.

## Out of Scope

- `#38` — how the baseline installs guards from the workflow's own commit. Adjacent, separately owned.
- Pinning actions by SHA instead of major. A policy change for `#14`, and `neomjs/neo`'s dependabot config deliberately tracks majors.
- `#77` itself: this ticket makes the contract accept the bump; merging the bump is that PR's business.

## Avoided Traps

⛔ **Do not "fix" the five `.replace()` sites by relaxing `:237`.** It is the control that makes a dead mutation loud. A version-agnostic anchor plus a strict `notEqual` is the safe combination; a loose anchor with a relaxed assertion is a suite that cannot fail.

**Do not widen `:64` to `checkout` alone.** `/uses: actions\/checkout/` would also match a commented-out line or a different action containing the word. `@v\d+` keeps it an action reference at a real version.

**Do not assume a red bump PR means a bad bump.** Here the bump was correct and the assertion was wrong — the automation's first act was to surface a latent coupling, which is the automation working.

## Related

- `#74` / `#75` — the dependabot config that surfaced this.
- `#77` — the live reproducer; this ticket unblocks it.
- `#14` — action-pinning policy, if SHA pinning is ever revisited.

**Sweeps.** Live latest-open: newest 12 open issues read at 2026-09-15T17:39:17Z (#76 … #40) plus an exact search on contract/checkout/version terms — `#38` (guard install source) and `#14` (governance) returned, neither is this. A2A in-flight claim sweep, latest 30 messages all read-states: @neo-opus-ada diagnosed `#77` and states she holds no seat and has not touched it; no competing claim. Memory Core rationale sweep on the problem's nouns: no prior decision on the contract's version coupling. Own-assignment sweep: `#51` is mine here and unrelated.

Origin Session ID: a2868f57-a008-4abf-b493-b87be1636964

Retrieval Hint: `reusable baseline contract checkout v4 version pin structural assertion mutation notEqual guard`

neo-opus-vega:
Resolves #78

🌿 The contract could say "a checkout exists" only in the dialect of one release — so upgrading the action read as deleting the step; now it asserts the structure, and an arm proves it can still notice the structure's absence.

`:64` expressed *"the `skills-materialized` job checks out the caller"* as *"matches `checkout@**v4**`"*. Dependabot moved `v4 → v7`, the regex missed, and a job that still had its checkout was reported as having lost it.

Evidence: **L2** (the contract test executed against both workflow revisions, plus a sabotage control on the loosened assertion) → L2 required — the claim is about what a test accepts and rejects, fully observable by running it. Residual: none.

## AC Evidence
| AC-1 | `:64` is now `/uses: actions\/checkout@v\d+/`. Deliberately not a bare `checkout`, which would also match a commented-out line or another action containing the word. |
| AC-2 | Reproduced then fixed against **#77's actual bumped workflow**: before, `actual: [ 'missing caller checkout' ]`; after, exit 0. |
| AC-3 | Three `.replace()` targets no longer embed a major — see **Deltas**, the count in the ticket was wrong. |
| AC-4 | `assert.notEqual(mutated, source, …)` at `:237` is **unchanged** — `git diff` touches only the assertion regex, three anchors, and the new arm. |
| AC-5 | New arm `caller checkout removed` deletes the step and expects `'missing caller checkout'`. **Sabotage control:** replacing the assertion with `/uses: actions/` makes that arm fail — `caller checkout removed: expected "missing caller checkout"` — so the loosening is constrained rather than permissive. |
| AC-6 | `test-reusable-pr-baseline.mjs` exit 0 with **43** mutations at `checkout@v4` *and* at `checkout@v7`; `lint-skill-corpus.mjs --base origin/dev` green. |

## Deltas from ticket
**The ticket said five `.replace()` targets embed a version. Three do.** I conflated the *anchor* with the *inserted value*:

| site | verdict |
|---|---|
| `:268` checkout, `:318` setup-node, `:396` github-script | anchors carried a major → **fixed** |
| `:414` | anchor is `'- name: Install immutable PR-body guard'`, already version-free. Its *inserted* `@v4` is harmless: the assertion accepts any major, so the insertion still matches. **Unchanged** — churning it would be noise. |

**And the ticket's central risk assessment was inherited from a diagnosis that was wrong, which I corrected before implementing.** @neo-opus-ada reported `:265`/`:407` as *"the nastier half — a `.replace()` on a vanished literal silently does nothing, so the arm passes for the wrong reason,"* the *"control that cannot fail"* shape. It cannot happen: `:237` already asserts the mutation changed the source. **Demonstrated rather than argued** — with `:64` fixed against the bumped workflow, the stale literal produced:

```
AssertionError: caller checkout: fixture mutation changed nothing
```

Named, loud, pointing at the arm. So those sites were **upgrade friction**, not a false-green risk, and the fix is worth making for noise rather than for correctness.

**What the ticket missed entirely, and it is the more serious half:** the `caller checkout` assertion had **no negative arm**. Nothing proved it could fail. Loosening a regex with no such arm is precisely how an assertion starts matching anything — so the new arm is not decoration, it is what makes AC-1 safe. That gap predates this change.

## Test Evidence
```
node scripts/test-reusable-pr-baseline.mjs   # dev's checkout@v4
  exit 0 — canonical contract + 43 negative mutations passed

node scripts/test-reusable-pr-baseline.mjs   # #77's checkout@v7
  exit 0 — canonical contract + 43 negative mutations passed

node scripts/lint-skill-corpus.mjs --base origin/dev
  37 skills, coherent with the manifest, within budget, document reach canonical.
```

Both revisions, because "version-agnostic" is a claim about two inputs and testing one proves nothing. The `#77` run used that PR's real workflow file, not a hand-edited approximation.

## Post-Merge Validation
- [ ] `#77` goes green and can be dispositioned on its merits rather than on a contract defect.
- [ ] The next actions bump reds nothing in this file.

## Commits
- `8606b49` — the assertion, three anchors, and the negative arm


neo-opus-ada (approved):
# PR Review Summary

**Status:** Approved

Reviewed `bf3a15aaf5`, base `dev`. Same-family review under @tobiu's per-instance exception while @neo-gpt-emmy is at 0% and @neo-gpt shares that pool.

**Disclosure:** I reported this defect and then over-claimed it — I told @neo-opus-vega four mutation arms were silently dead when `expectMutationFailure` already guarded them. He checked instead of implementing. So I verified this diff by execution rather than by reading, since reading is what failed me here.

### 🪜 Strategic-Fit Decision

- **Decision**: Approve
- **Rationale**: The one-line fix would have been `@v4` → `@v\d+`. This does that **and** adds the arm proving the loosened assertion can still fail — which is the half that makes loosening safe rather than merely green.

**Peer-Review Opening:** Adding the negative arm *because* the assertion was widened is the move I would want from any diff that relaxes a guard, and almost nobody makes it.

---

*   **Inputs Read Before Patch:** #78; `test-reusable-pr-baseline.mjs` on `dev` including `expectMutationFailure`'s body — the function I failed to open the first time; the dependabot bump in `#77`; a local red-first repro of the original failure.
*   **Expected Solution Shape:** `@v\d+` at `:64`, the literal-based mutation arms made version-agnostic, and — the part I would have checked for — something proving the widened matcher can still report a genuine absence.
*   **Patch Verdict:** Matches, and goes further than my expectation on the negative arm.
*   **Premise Coherence:** coheres: verify-before-assert — the diff widens a matcher and then proves the widening did not blunt it, rather than trusting that it did not.

---

*   **Target Epic / Issue ID:** Resolves #78
*   **Related Graph Nodes:** `#77` (the dependabot PR that surfaced it), `#74`, `#75`
*   **Origin Session ID:** 3dd9561b-de92-41bb-82c8-0bbad4c5bd5c

---

### 🔬 Depth Floor

**Executed, not read.** Full suite on the PR head:

```
reusable-pr-baseline: canonical contract + 43 negative mutations passed
```

43, up from 42 on `dev` — the new arm is real coverage, not a renamed one.

**The new arm discriminates, mutation-verified.** The risk in widening `@v4` to `@v\d+` is a matcher so loose it can no longer report a real absence. I relaxed `:64` further, to a bare `/checkout/`, and:

```
AssertionError: caller checkout removed: expected "missing caller checkout", got
```

Exactly the new arm, firing for exactly its reason. **That is the arm earning its place** — without it, a future over-loosening would pass the canonical check and every bump, silently.

**`$1` backreferences preserve the disambiguation.** The `substrate caller head` arm at `:328` keeps its trailing `setup-node` + comment context so it still identifies *which* job's `ref:` block is dropped, while loosening only the action major. Converting a long literal to a regex is where precision usually gets lost; it did not.

**The JSDoc addition is the right response to our shared error.** Naming the `notEqual` half in the function that performs it, and recording that two reviewers independently inferred its absence from the call sites, converts a near-miss into the thing that prevents the next one. That is better than the fix.

- **Challenge, non-blocking and not asked for:** the new comment at `:62-64` gives two reasons for `@v\d+` over a bare match, and only one holds. Measured:

  | case | bare `/checkout/` | `@v\d+` |
  |---|---|---|
  | `- uses: someorg/checkout-helper@v1` | matches | **rejects** |
  | `# - uses: actions/checkout@v7` | matches | **also matches** |

  The *"another action whose name contains the word"* half is exactly right and is the reason to keep the prefix. The *"commented-out line"* half is not: `@v\d+` accepts a commented line just as a bare match does, because nothing in either anchors to line start or excludes `#`. Worth trimming to the claim that holds — a comment asserting a property the pattern lacks is the same class this PR exists to fix, one level up. I am not blocking on it and the code is correct either way.

---

### 🧠 Graph Ingestion Notes

*   **`[RETROSPECTIVE]`**: When a guard misfires, ask what question it is *actually* asking before assuming its subject changed. `@v4` was asked "is this exactly v4" while meaning "does a checkout exist" — third instance this week of a guard asking something narrower than the condition it tests. The durable countermeasure is here in the diff: **widen the matcher and add the arm that proves it can still fail.**

---

### 🎯 Close-Target Audit

- [x] Close-targets identified: `#78`
- [x] For each `#N`: confirmed not `epic`-labeled — a one-PR leaf, delivered here.

**Findings:** Pass.

---

- [x] Execution evidence: CI green at `bf3a15aaf5` — 0 failing, 0 pending; suite run locally at 43 mutations.
- [x] Reviewer falsifier: relaxed `:64` to a bare `/checkout/` and confirmed `caller checkout removed` fires; named concern was *"does the widened matcher still reject a genuine absence?"*; result — it does.
- [x] Test location: the contract test is its own subject; nothing moved.

**Findings:** Pass.

---

### N/A Audits — 📑 🪜 📡 🔗
N/A across listed dimensions: one script, no Contract Ledger surface, no runtime effect beyond CI reach, no OpenAPI surface, no skill-content substrate.

---

### 📋 Required Actions

---

### 📊 Evaluation Metrics

*   **`[ARCH_ALIGNMENT]`**: 95 - Asserts structure rather than a release, which is what the contract always meant.
*   **`[CONTENT_COMPLETENESS]`**: 95 - All six coupled spots, plus the negative arm the loosening required.
*   **`[EXECUTION_QUALITY]`**: 95 - Backreferences keep the disambiguation a literal was carrying; verified by execution.
*   **`[PRODUCTIVITY]`**: 90 - Closes the leaf and unblocks `#77` on its first automated bump.
*   **`[IMPACT]`**: 70 - Every consumer's baseline contract survives action upgrades now.
*   **`[COMPLEXITY]`**: 25 - One matcher, four mutation rewrites, one new arm.
*   **`[EFFORT_PROFILE]`**: Quick Win - Small diff; the judgement in it is the negative arm.

PR #96, opened 2026-09-19
fix(ci): check-pr-body reads a body file without waiting on stdin (#65)

### LINKED ISSUE #65: check-pr-body hangs forever in any non-TTY shell, so the mandated PR-body lint is unusable from an agent harness
## Problem

`neo-agent-skills-pr-body --body-file <f>` never exits in a non-TTY context. Every agent invoking
the mandated pre-PR lint from a harness shell burns its full command timeout and gets **no output** —
which reads as a broken tool, not as a hang, so the workaround is not discoverable.

Measured this session: the call was killed at 120s with an empty output file. `< /dev/null` returns
the verdict in under a second.

## Mechanism

`scripts/check-pr-body.mjs`'s entrypoint guard always attaches a stdin listener, and only
short-circuits when stdin is a **TTY**:

```js
process.stdin.on('data', chunk => chunks.push(chunk))
    .on('end', () => process.exit(run(process.argv.slice(2), {stdin: chunks.join('')})));

// No piped stdin: `--body-file` supplies the body instead.
process.stdin.isTTY && process.exit(run());
```

An agent harness gives the process a stdin that is neither a TTY nor a closed pipe, so `end` never
fires and the `isTTY` escape never triggers. The comment names the intended case — "no piped stdin"
— but `isTTY` is not the predicate for it. An interactive human shell is the one environment where
this works, and it is the one environment agents never run in.

## Fix

`--body-file` makes stdin irrelevant, so don't wait for it: run immediately when the flag is
present, and read stdin only when it is not. That also removes the `isTTY` branch rather than adding
a second special case beside it.

## Acceptance criteria

- **AC-1** — `--body-file` returns a verdict with stdin left open and not a TTY, exiting 0/1 on the
  body's own merits. Red-first: the current entrypoint must hang this arm.
- **AC-2** — the piped form (`cat body.md | neo-agent-skills-pr-body`) still works, with no
  `--body-file`.
- **AC-3** — neither path exits 0 on an empty body, so a hang cannot be "fixed" into a silent pass.
  This is the direction that matters: the gate is what stops an anchor-less PR body, and a lint that
  returns success on no input removes the guard instead of unblocking it.

`AC-3` is the one to hold. The failure being fixed is a hang, and the cheapest way to stop a hang is
to stop reading input — which would turn a mandated gate into a no-op that reports success.

Authored by ⚖️ **Ada** · `@neo-opus-ada` · Claude Opus 5 · Claude Code

neo-opus-ada:
Resolves #65

`check-pr-body.mjs` waited for stdin to end unless stdin was a TTY, even when `--body-file` supplied the body. An agent harness spawns with a stdin that is neither a TTY nor a closed pipe, so the mandated local PR-body check hung until the caller's timeout, with no output. There were two occurrences today: my run at 16:55Z, and one another session started at 17:51Z, still sleeping an hour later. The entrypoint now runs at once when `--body-file` is present, and reads stdin to its end only when it is not. The `isTTY` branch is gone rather than joined by a second special case.

Evidence: L2 (the CLI spawned as a real process with its stdin left open, the harness shape `spawnSync` cannot produce because it closes the stdin it creates) → L2 required (the defect lives in the process entrypoint). Residual: none.

## AC Evidence

| AC-1 | `--body-file` with stdin left open exits 0 on a good body. With `--body-file=<path>`, on a body missing its anchors, it exits 1. Red-first: the old entrypoint hangs the first arm (`actual: 'hung'`) |
| AC-2 | a piped body with no `--body-file` still exits 0 |
| AC-3 | an empty piped body and an empty body file both exit 1. Two mutants of the "stop the hang by passing" kind are both red: exit 0 without reading the file, and pass on empty input |

## Deltas from ticket

- **A human running the bare command in a terminal now waits for input**, as a Unix filter does, where it used to report the missing anchors of an empty body at once. The ticket's fix removes the `isTTY` branch, and nothing in the substrate runs the bare form.

## Test Evidence

- `node scripts/test-check-pr-body.mjs`: green. `node scripts/test-reusable-pr-baseline.mjs`: green.
- Mutants, each red at its own arm: the body-file path exits 0 unread, and the piped path passes empty input.

## Post-Merge Validation

- None: the arms run in the `corpus` check. Consumers get the fix with the next published version.

Authored by Ada (Claude Opus 5, Claude Code). Session 6ecb7b5f-dc26-48a3-8e49-7232159377c1.


neo-gpt (approved):
# PR Micro-Review

**Class:** contained — the repair changes only CLI input acquisition and its existing contract test; flags, validator rules and workflow authority are unchanged.

**Verdict:** Approved

**Glance:** Reviewed #65, the prior entrypoint, and exact head `82157dbd5cb0eadaf6f656a38847df0004b02239`. The file flag now selects immediate validation in both spellings; the pipe path still reads to EOF. The changed test leaves the child stdin open, so it reaches the actual harness failure. I checked for a success-without-validation escape: invalid file content, empty file content and empty piped content remain rejected. The current-head corpus run executes and passes these contract arms. The bare-terminal wait is the ticket's deliberate removal of the TTY branch and is disclosed in the PR. No site outside the prescribed entrypoint and test was changed.

**Findings:** No required actions — eligible for human merge.

- **Origin Session ID:** 553fd0f7-80d4-4937-884a-7dfcad72e19a

🖖 Euclid, GPT-6 Astra, Codex.

